PR #9, opened 2025-09-03
browser profiles CLI commands

rgarcia:
- adds profile options to `browsers create`
- adds `profiles <create|get|delete|list|download>` commands


cmd/profiles.go
```diff
@@ -0,0 +1,302 @@
...
+					cp := pr
+					found = &cp
+					break
+				}
+			}
+		}
+		if found == nil {
+			pterm.Error.Printf("Profile '%s' not found\n", in.Identifier)
+			return nil
+		}
```
Sayan-, line 145: `List` over `Get` is a little surprising here tbh. This probably is negligible at first but would likely be worse in perf with deeper adoption

pkg/util/timefmt.go
```diff
@@ -0,0 +1,16 @@
+package util
+
+import "time"
+
+// DefaultTimeLayout is the standard layout used for displaying timestamps.
+// Includes the local timezone abbreviation to make it clear times are local.
+const DefaultTimeLayout = "2006-01-02 15:04:05 MST"
+
+// FormatLocal formats the provided time in the user's local timezone.
+// If the time is zero, it returns "-".
```
Sayan-, line 10: 👍

cmd/browsers.go
```diff
@@ -153,6 +164,21 @@ func (b BrowsersCmd) Create(ctx context.Context, in BrowsersCreateInput) error {
 		params.Headless = kernel.Opt(in.Headless.Value)
 	}
 
+	// Validate profile selection: at most one of profile-id or profile-name must be provided
+	if in.ProfileID != "" && in.ProfileName != "" {
+		pterm.Error.Println("must specify at most one of --profile-id or --profile-name")
+		return nil
```
Sayan-, line 170: return a non-nil error as well?

cmd/browsers.go
```diff
@@ -169,6 +195,13 @@ func (b BrowsersCmd) Create(ctx context.Context, in BrowsersCreateInput) error {
 	if browser.Persistence.ID != "" {
 		tableData = append(tableData, []string{"Persistent ID", browser.Persistence.ID})
 	}
+	if browser.Profile.ID != "" || browser.Profile.Name != "" {
+		profVal := browser.Profile.Name
+		if profVal == "" {
+			profVal = browser.Profile.ID
+		}
+		tableData = append(tableData, []string{"Profile", profVal})
+	}
```
Sayan-, line 204: for consistency should we be including `-` here?

Sayan- (changes requested):
Changes lgtm! Blocking until we've merged and updated the import!

Sayan- (approved):

PR #26, opened 2025-10-03
Truncate long columns instead of line wrapping

sjmiller609:
Detecting terminal size, use smart column truncation. Seek to avoid truncating shorter columns.

image

### DIFF
--- a/cmd/proxies/get.go
+++ b/cmd/proxies/get.go
@@ -38,6 +38,21 @@ func (p ProxyCmd) Get(ctx context.Context, in ProxyGetInput) error {
 	// Display type-specific config details
 	rows = append(rows, getProxyConfigRows(item)...)
 
+	// Display status with color
+	status := string(item.Status)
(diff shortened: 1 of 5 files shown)


cmd/proxies/get.go
```diff
@@ -38,6 +38,21 @@ func (p ProxyCmd) Get(ctx context.Context, in ProxyGetInput) error {
...
+	} else if status == "available" {
+		status = pterm.Green(status)
+	} else if status == "unavailable" {
+		status = pterm.Red(status)
+	}
+	rows = append(rows, []string{"Status", status})
+
+	// Display last checked timestamp
+	lastChecked := util.FormatLocal(item.LastChecked)
+	rows = append(rows, []string{"Last Checked", lastChecked})
```
sjmiller609, line 54: noticed forgot to include status info in `kernel proxies get...` in a recent pr

rgarcia (approved):

PR #134, opened 2026-03-13
Add Tzafon Northstar computer use template

mttias:
Adds TypeScript and Python templates for Tzafon's Northstar CUA Fast model using Kernel Computer Controls API. Tested end-to-end on Kernel.

Attaching video of the following task: "go to magnitasks.com and click on "Tasks" in the sidebar. Then, drag the task cards from "To Do" to "Done" column without clicking into them.".
https://github.com/user-attachments/assets/b42f816f-c0b8-4b32-a6c0-0fbf2c8d8f15

---

> [!NOTE]
> **Low Risk**
> Low risk because changes are additive and mostly isolated to new template scaffolding, with only a small update to template selection/sorting and sample commands.
> 
> **Overview**
> Adds a new `tzafon` template option to `pkg/create/templates.go`, including display metadata, sort priority, and deploy/invoke sample commands for both TypeScript and Python.
> 
> Introduces new `pkg/templates/{typescript,python}/tzafon` reference apps that run a Northstar CUA loop via Tzafon Lightcone, including browser-session management (optional replay recording), computer-control tool implementations (click/type/key/scroll/drag), and corresponding `.env.example`/README/dependency configs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 678ebf68997cb9fc8788f3c686fb4f373a3e3cb8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/templates/typescript/tzafon-computer-use/tools/computer.ts
```diff
@@ -0,0 +1,178 @@
...
+export class ToolError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'ToolError';
+  }
+}
+
+const MODIFIER_MAP: Record<string, string> = {
+  Control: 'Ctrl',
+  Enter: 'Return',
```
dprevoznik: **dp-cursor-review-analysis**

The `MODIFIER_MAP` only handles `"Enter"` and `"Control"` (capitalized), but the Tzafon model sends `"enter"` (lowercase). This causes `pressKey` to receive the literal string `"enter"` instead of the X11 keysym `"Return"` that Kernel's API expects. Fixed by adding lowercase variants to the map.

pkg/templates/typescript/tzafon-computer-use/tools/computer.ts
```diff
@@ -0,0 +1,178 @@
...
+
+      case 'navigate':
+        await this.kernel.browsers.playwright.execute(this.sessionId, {
+          code: `await page.goto(${JSON.stringify(action.url)})`,
+        });
+        break;
+
+      case 'drag':
+        await this.kernel.browsers.computer.dragMouse(this.sessionId, {
+          path: [
```
dprevoznik: **dp-cursor-review-analysis**

Drag actions from the Tzafon model are incomplete. The model emits `{"type":"drag","x":470,"y":344}` with only a start point — no `end_x`/`end_y` or `x2`/`y2`. This causes `dragMouse` to receive `NaN` for the destination, so the drag silently does nothing. This is a Tzafon model-side limitation; the template code handles drag correctly when both start and end coordinates are provided.

pkg/templates/python/tzafon-computer-use/tools/computer.py
```diff
@@ -0,0 +1,144 @@
...
+import asyncio
+import base64
+import json
+from typing import Any
+
+from kernel import Kernel
+
+from .base import ToolError
+
+MODIFIER_MAP = {"Control": "Ctrl", "Enter": "Return"}
```
dprevoznik: **dp-cursor-review-analysis**

Same issue as TypeScript — `MODIFIER_MAP` only handles `"Enter"` and `"Control"` (capitalized), but the Tzafon model sends `"enter"` (lowercase). This causes `press_key` to receive `"enter"` instead of `"Return"`. Fixed by adding lowercase variants to the map.

Sayan- (approved):
couple of small q's. overall looks good!

pkg/templates/typescript/tzafon-computer-use/tools/computer.ts
mttias: Addressed in recent PR + model update.

pkg/templates/python/tzafon-computer-use/tools/computer.py
```diff
@@ -0,0 +1,144 @@
...
+                self.session_id,
+                x=clamp(self._center_x() if action.x is None else action.x, self.width),
+                y=clamp(self._center_y() if action.y is None else action.y, self.height),
+                delta_x=0 if action.scroll_x is None else action.scroll_x,
+            )
+        elif t == "navigate":
+            self.kernel.browsers.playwright.execute(
+                self.session_id, code=f"await page.goto({json.dumps(action.url)})",
+            )
+        elif t == "drag":
```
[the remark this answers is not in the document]
mttias: Addressed in recent PR + model update.

pkg/templates/python/tzafon-computer-use/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[project]
+name = "python-tzafon-cua"
+version = "0.1.0"
+description = "Kernel reference app for Tzafon Northstar Computer Use"
+requires-python = ">=3.11"
+dependencies = [
+    "kernel>=0.35.0",
+    "tzafon",
```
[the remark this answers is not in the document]
mttias: Addressed in recent PR update.

pkg/templates/python/tzafon-computer-use/tools/computer.py
```diff
@@ -0,0 +1,159 @@
...
+                self.session_id,
+                path=[
+                    [clamp(x1, self.width), clamp(y1, self.height)],
+                    [clamp(x2, self.width), clamp(y2, self.height)],
+                ],
+            )
+        elif t == "wait":
+            await asyncio.sleep(2)
+        else:
+            raise ToolError(f"Unknown action type: {t}")
```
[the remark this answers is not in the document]
mttias: Resolved — we've moved from the computer_call protocol to explicit function tools. click now takes an optional button param (left/right). README updated to reflect the current tool set.

mttias:
Made a recent update that fixed some of the minor issues such as drag n drop.

dprevoznik:
## Suggestion: Add a system prompt via `instructions`

The Lightcone Responses API supports an `instructions` field that acts as a system prompt ([docs](https://docs.lightcone.ai/guides/responses-api/index.md#system-instructions)). The tzafon template currently doesn't use it — the task is passed directly as the first user message with no system-level guidance.

The other templates (Anthropic, Gemini) all include system prompts with environment context and behavioral hints. For reference, here's what the Anthropic template uses:

```
<SYSTEM_CAPABILITY>
* You are utilising an Ubuntu virtual machine with internet access.
* When you connect to the display, CHROMIUM IS ALREADY OPEN. The url bar is not visible but it is there.
* If you need to navigate to a new page, use ctrl+l to focus the url bar and then enter the url.
* You won't be able to see the url bar from the screenshot but ctrl-l still works.
* As the initial step click on the search bar.
* When viewing a page it can be helpful to zoom out so that you can see everything on the page.
* Either that, or make sure you scroll down to see everything before deciding something isn't available.
* Scroll action: scroll_amount and the tool result are in wheel units (not pixels).
* When using your computer function calls, they take a while to run and send back to you.
* Where possible/feasible, try to chain multiple of these calls all into one function calls request.
* The current date is <date>.
* After each step, take a screenshot and carefully evaluate if you have achieved the right outcome.
* Explicitly show your thinking: "I have evaluated step X..." If not correct, try again.
* Only when you confirm a step was executed correctly should you move on to the next one.
</SYSTEM_CAPABILITY>

<IMPORTANT>
* When using Chromium, if a startup wizard appears, IGNORE IT. Do not even click "skip this step".
* Instead, click on the search bar on the center of the screen where it says "Search or enter address", and enter the appropriate search term or URL there.
</IMPORTANT>
```

**Proposed addition** — add an `instructions` field to the `responses.create()` call in both TS and Python loops. Things to include:

- **Environment context**: Chromium is already open, url bar behavior, ctrl+l navigation
- **Scrolling guidance**: Prefer `key("pagedown")` / `key("pageup")` for full-page scrolling (the key map already supports these) — the `scroll` tool is mouse-wheel-style and can be unreliable for large page traversals
- **Date**: Current date for time-sensitive tasks
- **Step verification**: Evaluate each step before moving on
- **Chromium startup wizard**: Ignore it if it appears

Some of the Anthropic hints may not apply directly (e.g., scroll_amount units, chaining calls) since Northstar uses explicit function tools rather than Anthropic's `computer_use` tool type. We should adapt rather than copy verbatim.

pkg/templates/typescript/tzafon-computer-use/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../tsconfig.base.json",
```
dprevoznik: this `tsconfig.json` extends `../tsconfig.base.json`, but `kernel create` only copies files from inside the template directory — so users who scaffold this template won't have the base config. the tsconfig ends up with no compiler options applied (no `strict`, no `moduleResolution`, etc.). you can see the warning during deploy: `Cannot find base config file "../tsconfig.base.json"`.

the other templates (sample-app, anthropic-computer-use, stagehand, etc.) inline the compiler options directly. might be worth doing the same here.

pkg/templates/typescript/tzafon/loop.ts
```diff
@@ -0,0 +1,244 @@
...
+
+    // Prevent unbounded payload growth — keep the task prompt + recent history
+    if (items.length > 30) {
+      items.splice(2, items.length - 22);
+    }
+
+    resp = await tzafon.responses.create({
+      model,
+      input: items,
+      tools: TOOLS,
```
dprevoznik, line 186: the `TOOLS` array causes a TS compilation error because the `@tzafon/lightcone` SDK's `Tool` type only allows `type: "computer_use" | "computer_use_preview"`, not `"function"`. the API accepts function tools fine at runtime, but the type definitions need updating on the tzafon team side — users who scaffold this template will see type errors in their IDE.

Sayan- (approved):
Thanks for this contribution!

dprevoznik (approved):
LGTM!

PR #141, opened 2026-04-02
Add singular alias for auth connections subcommand

sjmiller609:
## Summary
- Add `connection` as an alias for the `connections` subcommand under `kernel auth`, so `kernel auth connection list` works alongside `kernel auth connections list`
- This was the only resource command missing a singular/plural alias — all others already had both forms

## Context
Every other resource command in the CLI already supports both singular and plural:
- `browsers` / `browser`
- `browser-pools` / `browser-pool` / `pool` / `pools`
- `app` / `apps`
- `profiles` / `profile`
- `proxies` / `proxy`
- `extensions` / `extension`
- `credentials` / `credential` / `creds` / `cred`
- `credential-providers` / `credential-provider` / `cred-providers` / `cred-provider`
- `logs` / `log`

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./cmd/...` passes

### DIFF
--- a/cmd/auth_connections.go
+++ b/cmd/auth_connections.go
@@ -711,9 +711,10 @@ func (c AuthConnectionCmd) Follow(ctx context.Context, in AuthConnectionFollowIn
 // --- Cobra wiring ---
 
 var authConnectionsCmd = &cobra.Command{
-	Use:   "connections",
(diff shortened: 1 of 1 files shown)


sjmiller609:
wrong repo

PR #143, opened 2026-04-08
Add unified CUA template with multi-provider fallback

masnwilliams:
## Summary

- Adds a new `cua` template (TypeScript + Python) that consolidates the separate `anthropic-computer-use`, `openai-computer-use`, and `gemini-computer-use` templates into a single multi-provider template
- Provider selection via `CUA_PROVIDER` env var, with automatic fallback via `CUA_FALLBACK_PROVIDERS`
- Each provider adapter is self-contained with full agent loop implementation
- Shared browser session lifecycle with replay recording support
- Registered as a new template in `templates.go` for both TypeScript and Python

## Structure

```
pkg/templates/{typescript,python}/cua/
  index.ts / main.py          — Kernel app entrypoint
  session.ts / session.py     — Browser session lifecycle
  providers/
    index.ts / __init__.py    — Provider factory + fallback
    anthropic.ts / .py        — Anthropic Claude adapter
    openai.ts / .py           — OpenAI GPT adapter
    gemini.ts / .py           — Google Gemini adapter
```

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./pkg/create/...` passes
- [ ] `kernel create` shows "Unified CUA" template for both TS and Python
- [ ] Deploy TS template with Anthropic provider and run a task
- [ ] Deploy TS template with OpenAI provider and run a task
- [ ] Deploy TS template with Gemini provider and run a task
- [ ] Test fallback by setting an invalid primary key with valid fallback
- [ ] Repeat above for Python template

---

> [!NOTE]
> **Medium Risk**
> Adds a sizable new template with provider-selection/fallback logic and browser session lifecycle/replay handling, which may affect new-user flows and external API integrations. Existing templates are mostly untouched aside from template registry/sorting updates.
> 
> **Overview**
> Adds a new `cua` (“Unified CUA”) template for **both TypeScript and Python** that runs a computer-use agent against Anthropic/OpenAI/Gemini with `CUA_PROVIDER` selection and optional `CUA_FALLBACK_PROVIDERS` automatic fallback.
> 
> Registers the new template in `pkg/create/templates.go` (including deploy/invoke samples and template ordering), and introduces new template projects under `pkg/templates/{typescript,python}/cua` with provider-specific adapters, a shared browser session manager (including optional replay recording), and accompanying `.env.example`/README/dependency files.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 51b69fb0ae1bd5ed68236e1c8832b2541ec7f76e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


pkg/templates/typescript/cua/providers/index.ts
```diff
@@ -0,0 +1,114 @@
...
+
+import type { Kernel } from '@onkernel/sdk';
+import { AnthropicProvider } from './anthropic';
+import { OpenAIProvider } from './openai';
+import { GeminiProvider } from './gemini';
+
+// Shared interface every provider adapter must implement.
+export interface TaskOptions {
+  query: string;
+  model?: string;
```
dprevoznik, line 19: Each provider hardcodes model-specific API features that will break if you swap in a different model. Sharing Cursor's analysis, which is pretty aligned with what i experienced building other templates. I think it would be smart to at least have in the template which models from which providers are compatible. I don't think we need to lock it down to specific models, particularly if we have defaults.

Image


**Cursor Reco**
The model field should either be removed from the public payload (keep it as an env var only) or the template should validate model compatibility per provider before calling the API. At minimum, the README and input description should warn that only specific computer-use-capable models work. Right now a user seeing that free-text field in the dashboard will absolutely try claude-haiku-3-5 or gpt-4o and get a confusing 400 error.

pkg/templates/typescript/cua/providers/gemini.ts
```diff
@@ -0,0 +1,247 @@
...
+          const x = this.denormalize(args.x, width);
+          const y = this.denormalize(args.y, height);
+          await computer.clickMouse(sessionId, { x, y });
+          if (args.text) {
+            await computer.typeText(sessionId, { text: args.text });
+          }
+          break;
+        }
+        case 'scroll_document':
+        case 'scroll_at': {
```
dprevoznik, line 183: The Gemini provider in both ts-cua and python-cua is not likely optimized for scroll behavior. In the standalone templates, we have logic in place (though not perfect) to handle the fact that Gemini reports back a magnitude as the value in pixels.

Cursor summary of issue: 
`It's missing the magnitude ÷ 60 pixel-to-notch conversion and the max(1, min(17, ...)) clamp. When Gemini asks to scroll with its default magnitude (~400 pixels), the unified template will fire 400 wheel notches instead of 7. Both TS and Python have the same bug — they're consistent with each other, but both wrong compared to the standalone template.`

Getting API exhaustion errors with gemini right now, but wanted to surface this. I put in the % 60 and clamp logic in the standalone templates. It likely could be improved

dprevoznik: The behavior in anthropic and openai providers matches the standalone templates right now.

dprevoznik (commented):
Left some comments. All three providers working on both ts and python versions. Though Gemini api exhaustion error made it so I couldn't test the scroll logic comment I made.

dprevoznik (approved):
Other models look good. Everything working. 

Gemini feedback: 

System-prompt level Scroll behavior change makes sense. We may still want to change the default scroll action handler logic as I suggested since if the model chooses to use regular scroll it will be large jumps currently. Thoughts?

PR #146, opened 2026-04-09
feat: add `kernel browsers curl` command

rgarcia:
## Summary

Adds a curl-like `kernel browsers curl` command for making HTTP requests through a browser session's Chrome network stack. Requests inherit the browser's TLS fingerprint, cookies, proxy configuration, and browser headers, and responses are streamed directly through the SDK browser HTTP client wrapper around `/curl/raw`.

`/curl/raw` follows redirects automatically in Chromium, so this command documents that behavior rather than exposing `-L`.

## Usage

```bash
# Simple GET
kernel browsers curl $ID https://example.com

# POST JSON through the browser session
kernel browsers curl $ID https://api.example.com \
  -H "Content-Type: application/json" \
  -d '{"key":"val"}'

# Include response headers and write headers + body to a file
kernel browsers curl $ID -i -o page.html https://example.com

# Dump headers separately from the response body
kernel browsers curl $ID -D headers.txt -o body.html https://example.com

# Fetch headers only
kernel browsers curl $ID -I https://example.com

# Fail on HTTP errors without printing the response body
kernel browsers curl $ID -f https://example.com/missing

# Print curl-style metrics after completion
kernel browsers curl $ID -o body.html -w 'status=%{http_code} bytes=%{size_download}\n' https://example.com
```

## Flags

| Flag | Description |
|------|-------------|
| `-X, --request` | HTTP method (default: GET; defaults to POST when `--data` is set) |
| `-H, --header` | HTTP header, repeatable (`"Key: Value"` format) |
| `-d, --data` | Request body |
| `--data-file` | Read request body from file |
| `--max-time` | Maximum time allowed for the request in seconds (default: 30) |
| `-o, --output` | Write response body to file |
| `-I, --head` | Fetch headers only |
| `-i, --include` | Include response headers in output |
| `-D, --dump-header` | Write received headers to file (use `-` for stdout) |
| `-w, --write-out` | Output text after completion; supports `%{http_code}`, `%{response_code}`, `%{time_total}`, `%{size_download}` |
| `-f, --fail` | Fail with no body output on HTTP errors |
| `-s, --silent` | Suppress progress output |

## Implementation notes

- Uses `github.com/kernel/kernel-go-sdk v0.51.0` and `Browsers.HTTPClient(sessionID)` to route requests through `/curl/raw`.
- Always streams the proxied response directly; there is no JSON envelope mode.
- Redirects are followed automatically by Chromium and currently cannot be disabled.
- Root command error handling now writes CLI errors to stderr, which keeps `--fail` and timeout output closer to curl behavior.

## Test plan

- [x] `make test && make build`
- [x] Production smoke: create browser, run `kernel browsers curl <id> https://example.com`, delete browser
- [x] Compare actual `curl` vs `kernel browsers curl` for body output, `-i`, `-o`, `-i -o`, and `-d`
- [x] Compare `-I`, `-D`, `-w`, `-f`, and `--max-time`
- [x] Verify `/curl/raw` follows redirects automatically while plain `curl` does not without `-L`

---

> [!NOTE]
> **Medium Risk**
> Introduces new HTTP execution paths and modifies global error output behavior; while scoped to a new subcommand, it changes stdout/stderr expectations and request/response streaming semantics.
> 
> **Overview**
> Adds a new `kernel browsers curl <session-id> <url>` command that makes HTTP requests through the browser session’s SDK-provided `HTTPClient`, supporting curl-like flags for method/headers/body (incl. `--data-file`), timeouts, output-to-file, header inclusion/dumping, `--write-out` metrics, `--fail`, and `--silent`.
> 
> Updates root error handling to support “silent” command errors (suppress diagnostics while still exiting non-zero) and to ensure CLI errors are written to fang’s error stream (stderr), keeping stdout clean for response bodies. Documentation and examples in `README.md` are updated, and new unit tests cover request shaping, output/header behaviors, write-out expansion, fail mode, and silent error wrapping.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db97766dfc73fb351529e4a8f38205014dfdd71d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/browsers.go
+++ b/cmd/browsers.go
@@ -14,6 +14,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/kernel/cli/pkg/table"
 	"github.com/kernel/cli/pkg/util"
@@ -36,6 +37,7 @@ type BrowsersService interface {
 	Update(ctx context.Context, id string, body kernel.BrowserUpdateParams, opts ...option.RequestOption) (res *kernel.BrowserUpdateResponse, err error)
 	Delete(ctx context.Context, body kernel.BrowserDeleteParams, opts ...option.RequestOption) (err error)
 	DeleteByID(ctx context.Context, id string, opts ...option.RequestOption) (err error)
+	HTTPClient(id string, opts ...option.RequestOption) (*http.Client, error)
 	LoadExtensions(ctx context.Context, id string, body kernel.BrowserLoadExtensionsParams, opts ...option.RequestOption) (err error)
 }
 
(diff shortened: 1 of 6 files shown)


cmd/browsers.go
```diff
@@ -3256,6 +3283,281 @@ func runBrowsersComputerWriteClipboard(cmd *cobra.Command, args []string) error
...
+
+	var bodyReader io.Reader
+	if body != "" {
+		bodyReader = strings.NewReader(body)
+	}
+
+	// Seed the SDK's browser route cache before constructing the raw curl client.
+	if _, err := b.browsers.Get(ctx, in.Identifier, kernel.BrowserGetParams{}); err != nil {
+		return curlError(in, util.CleanedUpSdkError{Err: err})
+	}
```
rgarcia, line 3428: we might want to consider a local disk cache of this since this is super inefficient

cmd/browsers.go
```diff
@@ -3256,6 +3283,281 @@ func runBrowsersComputerWriteClipboard(cmd *cobra.Command, args []string) error
...
+
+	httpClient, err := b.browsers.HTTPClient(in.Identifier)
+	if err != nil {
+		return curlError(in, util.CleanedUpSdkError{Err: err})
+	}
+	if in.MaxTime > 0 {
+		httpClient.Timeout = in.MaxTime
+	}
+
+	req, err := http.NewRequestWithContext(ctx, method, in.URL, bodyReader)
```
hiroTamada, line 3438: can we verify that `/curl/raw` doesn't forward Go's default `User-Agent` / `Accept-Encoding` as browser request headers when the user didn't specify them?

The CLI builds a normal Go `http.Request` and sends it through the SDK `HTTPClient`; Go's transport can add defaults like `User-Agent: Go-http-client/1.1`, and the raw endpoint appears to forward non-hop-by-hop headers into the Chromium forward proxy. If the Chromium proxy honors those forwarded headers, this could override the browser's UA/default headers even though the command is documented as inheriting browser headers.

hiroTamada (approved):

cmd/browsers.go
rgarcia: addressing upstream in chromium

PR #147, opened 2026-04-10
feat(cli): add projects support with --project flag and name resolution

hiroTamada:
## Summary

- Adds a global `--project` persistent flag (and `KERNEL_PROJECT_ID` env var) that injects the `X-Kernel-Project-Id` header to scope all API requests to a specific project
- The flag accepts **either a project ID or a project name** — names are resolved via the projects list endpoint with case-insensitive matching
- Adds `kernel projects` subcommands: `list`, `create`, `get`, `delete`, `get-limits`, `set-limits`
- Upgrades `kernel-go-sdk` to v0.48.0 for project endpoint support (also fixes `ProxyService.Check` signature change)

## How it works

When `--project` is provided:
1. If the value looks like a cuid2 ID (24 lowercase alphanumeric chars), it's used directly as the project ID header
2. Otherwise, the CLI authenticates first, calls `GET /projects` to list all projects, finds a case-insensitive name match, and injects the resolved ID
3. Clear errors for: no match found, multiple matches (ambiguous name)

## Test plan

Tested against staging with a project-scoped API key:

- [x] `--project "cli-test-project"` (by name) — resolves correctly
- [x] `--project "CLI-TEST-PROJECT"` (case-insensitive) — resolves correctly
- [x] `--project "g9vcya6unur84k70n0u8s9p9"` (by ID) — works directly
- [x] `--project "nonexistent-project"` — clear error message
- [x] `KERNEL_PROJECT_ID="cli-test-project"` env var with name — resolves correctly
- [x] No `--project` with scoped API key — auto-scopes via server middleware
- [x] `kernel projects list/create/get/delete/get-limits/set-limits` — all functional

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Introduces a new global request-scoping mechanism via the `X-Kernel-Project-Id` header and adds project CRUD/limits commands, which can affect all API calls when enabled. Also bumps the Kernel SDK and adjusts proxy health-check calls to match the new SDK signature.
> 
> **Overview**
> Adds first-class **project management** to the CLI via a new `projects` command group (`list`, `create`, `get`, `delete`, plus `limits get/set` with optional `--output json`) and validation for limit updates (rejects negative values, treats unset vs set via `Int64Flag`).
> 
> Introduces a global `--project` flag (or `KERNEL_PROJECT` env var) that injects `X-Kernel-Project-Id` into all authenticated requests, plus project *name → ID* resolution via paginated project listing with clear errors for no/ambiguous matches.
> 
> Updates CUID2 validation to require a leading letter, upgrades `github.com/kernel/kernel-go-sdk` to `v0.48.0`, and adapts proxy health-check wiring/tests to the SDK’s updated `Check` signature.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 748bef4a785a78f3265bdcc915c0bc45d0c8b2aa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


cmd/root.go
```diff
@@ -122,12 +124,41 @@ func init() {
...
 
-		// Get authenticated client with OAuth tokens or API key fallback
-		client, err := auth.GetAuthenticatedClient(option.WithHeader("X-Kernel-Cli-Version", metadata.Version))
+		clientOpts := []option.RequestOption{
+			option.WithHeader("X-Kernel-Cli-Version", metadata.Version),
+		}
+
+		projectVal, _ := cmd.Flags().GetString("project")
+		if projectVal == "" {
+			projectVal = os.Getenv("KERNEL_PROJECT_ID")
```
rgarcia: nit: since this now accepts either a project name or ID, should this env var be called `KERNEL_PROJECT` to match `--project` and better signal the flexibility?

cmd/root.go
```diff
@@ -223,6 +255,41 @@ func isUsageError(err error) bool {
...
+// looksLikeCUID returns true if s matches the cuid2 format used for resource IDs.
+// Delegates to the shared cuidRegex defined in browsers.go.
+func looksLikeCUID(s string) bool {
+	return cuidRegex.MatchString(s)
+}
+
+// resolveProjectByName lists the caller's projects and returns the ID of the
+// one whose name matches (case-insensitive). Returns an error if no match or
+// multiple matches are found.
+func resolveProjectByName(ctx context.Context, client kernel.Client, name string) (string, error) {
```
rgarcia: curious why we need client-side name resolution here. if the API/SDK already has a get-by-id-or-name path under the hood, might be worth delegating to that instead of listing + matching locally

cmd/projects.go
```diff
@@ -0,0 +1,240 @@
...
+	"fmt"
+
+	"github.com/kernel/kernel-go-sdk"
+	"github.com/kernel/kernel-go-sdk/packages/param"
+
+	"github.com/pterm/pterm"
+	"github.com/spf13/cobra"
+)
+
+var projectsCmd = &cobra.Command{
```
rgarcia: question on code organization: browsers uses a cobra-independent struct + typed input pattern (`BrowsersCmd`, `BrowsersListInput`, etc.), which seems to make the logic easier to test and keeps flag parsing separate from API behavior/output formatting. since projects is already a multi-command surface, would it be worth following that pattern here too with a `ProjectsCmd`?

cmd/projects.go
```diff
@@ -0,0 +1,240 @@
...
+			return nil
+		}
+
+		pterm.Success.Printf("Deleted project: %s\n", projectID)
+		return nil
+	},
+}
+
+var projectsLimitsGetCmd = &cobra.Command{
+	Use:   "get-limits <id-or-name>",
```
rgarcia: question on command shape: these `get-limits` / `set-limits` verbs feel a little inconsistent with the rest of the cli. most resources use `list`/`get`/`create`/`update`/`delete`, and when an area grows we seem to introduce a nested noun subtree like `browsers fs ...` or `browsers computer ...`. if project admin actions are going to expand, would `projects limits get` / `projects limits set` be a better fit?

cmd/projects.go
```diff
@@ -0,0 +1,240 @@
...
+			return err
+		}
+
+		limits, err := client.Projects.Limits.Get(ctx, projectID)
+		if err != nil {
+			pterm.Error.Println("Failed to get project limits:", err)
+			return nil
+		}
+
+		out, _ := json.MarshalIndent(limits, "", "  ")
```
rgarcia: looks like these limits commands always print raw json. the rest of the cli seems to reserve machine-readable output for an explicit `--output json` / `-o json` flag and otherwise uses human-friendly tables/messages. might be worth making projects follow that same convention for consistency

rgarcia (dismissed):

masnwilliams (approved):
lgtm — all review comments addressed. typed handler pattern, nested limits subcommands, human-readable output, and KERNEL_PROJECT precedence all look good.

cmd/root.go
```diff
@@ -122,8 +134,18 @@ func init() {
...
+		clientOpts := []option.RequestOption{
+			option.WithHeader("X-Kernel-Cli-Version", metadata.Version),
+		}
+
+		projectVal, _ := cmd.Flags().GetString("project")
+		projectVal = resolveProjectSelection(projectVal)
+
+		if projectVal != "" {
+			clientOpts = append(clientOpts, option.WithHeader("X-Kernel-Project-Id", projectVal))
+		}
```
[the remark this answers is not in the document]
sjmiller609: seems worthy

sjmiller609 (approved):

masnwilliams:
Pushed a commit dropping the `KERNEL_PROJECT_ID` fallback per the docs thread — projects is brand new, there's nothing to preserve.

One other thing worth a quick check before merge: the PR description says the `--project` global flag resolves names → IDs client-side ("If the value looks like a cuid2 ID… it's used directly as the project ID header. Otherwise, the CLI authenticates first, calls `GET /projects`… and injects the resolved ID"), but in `cmd/root.go` the `PersistentPreRunE` passes `projectVal` straight into the `X-Kernel-Project-Id` header with no call to `resolveProjectByName`:

```go
projectVal, _ := cmd.Flags().GetString("project")
projectVal = resolveProjectSelection(projectVal)

if projectVal != "" {
    clientOpts = append(clientOpts, option.WithHeader("X-Kernel-Project-Id", projectVal))
}
```

`resolveProjectByName` is only wired into `resolveProjectArg` in `cmd/projects.go` for positional args on `projects get`/`delete`/`limits`.

So either:
1. The server resolves names from the header too (in which case the description is fine, but the cuid2 regex check in the description is misleading since it doesn't run), or
2. `kernel browsers list --project staging` sends `staging` as the header literal and the API rejects it.

Your test plan shows `--project "cli-test-project"` passing — is that because the server accepts names in `X-Kernel-Project-Id`, or was it happening to hit the positional-arg path of a `projects` subcommand? Worth adding `resolveProjectByName` in the `PersistentPreRunE` path if it's #2.

PR #180, opened 2026-06-09
cli: add custom chrome policy flags to browser and pool commands

rgarcia:
## Summary

Adds CLI support for setting a custom Chrome enterprise policy (the API's `chrome_policy` field, already supported by the Go SDK) on three commands:

- `kernel browsers create`
- `kernel browser-pools create`
- `kernel browser-pools update`

Each gains two mutually-exclusive flags:

- `--chrome-policy '<json>'` — an inline JSON object
- `--chrome-policy-file <path>` — read the object from a file (`-` reads stdin)

This mirrors the existing `--payload` / `--payload-file` and `--data` / `--data-file` conventions.

```bash
kernel browsers create --chrome-policy '{"BookmarkBarEnabled": false}'
kernel browsers create --chrome-policy-file policy.json
cat policy.json | kernel browser-pools update my-pool --chrome-policy-file -
```

## Design notes

- A shared `parseChromePolicy` helper parses into `map[string]any`, so non-object JSON (arrays, scalars, `null`) is rejected with a clear error before any request is sent.
- `chrome_policy` uses `omitzero`, which only drops a **nil** map — a non-nil empty `{}` would serialize as `"chrome_policy":{}`. Assignment is therefore guarded by `len > 0`, so an empty object is treated as "unset".
- On `browser-pools update`, an empty `{}` cannot clear an existing policy (omitzero strips it before it reaches the server). Rather than silently no-op, this surfaces a warning in human output (suppressed under `-o json` to keep stdout valid).
- Kernel-managed policies (extensions, proxy, automation) are enforced **server-side**. The CLI forwards the object verbatim and surfaces the API's error, rather than duplicating a blocklist that would drift.

## Scope

- `browsers update` and `invoke` are intentionally excluded: their SDK params structs (`BrowserUpdateParams`, `InvocationNewParams`) carry no `chrome_policy` field, so there's nothing to wire until the SDK is regenerated.
- The policy echoes back via `--output json` with no printer changes (response structs already carry the field).

## Test plan

- [x] Unit tests for all three commands: inline policy applied, empty `{}` omitted (asserted at both the Go field and the serialized JSON), invalid JSON / `null` rejected, parser covers file + stdin + whitespace + missing-file, and the pool-lease conflict contract is locked (`--chrome-policy` stays out of the allowed-flag set).
- [x] `make test` (go vet + go test ./...) passes; gofmt clean.
- [x] E2E against a live environment via the built binary: inline/file/stdin create echo the policy; pool create + get persist it; pool update replaces it and an empty `{}` no-ops without clearing; invalid/null/mutual-exclusion exit non-zero before any API call; and the server correctly rejects `ProxyMode` / `ExtensionInstallForcelist` with the CLI surfacing the error.

## Note (pre-existing, unrelated)

When a browser-config flag is combined with `--pool-id`/`--pool-name`, the existing conflict-confirmation prompt (`pterm.DefaultInteractiveConfirm`) does not treat EOF / non-TTY stdin as "no", so it blocks in a non-interactive pipeline. This predates this change and affects all config flags equally; flagging it as a possible follow-up.

---

> [!NOTE]
> **Low Risk**
> CLI-only feature forwarding JSON to existing API fields; validation is local and server enforces blocked policies. No auth or data-path changes.
> 
> **Overview**
> Adds **`--chrome-policy`** and **`--chrome-policy-file`** (mutually exclusive, `-` for stdin) to `kernel browsers create` and `kernel browser-pools create` / `update`, wiring the API’s `chrome_policy` field through a shared **`parseChromePolicy`** helper.
> 
> Policies are only sent when the parsed object is non-empty (`len > 0`), so `{}` is not serialized on create. On **pool update**, passing `{}` does not clear an existing policy; the CLI warns in human output and stays silent with `-o json`. **`poolLeaseAllowedFlags`** is extracted so `--chrome-policy` is **not** treated as allowed alongside `--pool-id`/`--pool-name`, surfacing the existing pool-lease conflict warning instead of ignoring the flag.
> 
> README documents flags, examples, and the pool-update clearing limitation. Unit tests cover parsing, SDK params, JSON omission, warnings, and pool-lease behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a470a3abdbd12ddc372196c38fb6297adb2259d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,8 @@ Commands with JSON output support:
   - `--telemetry=all` - Enable telemetry for all categories
   - `--telemetry=off` - Disable telemetry
   - `--telemetry=<list>` - Per-category config, e.g. `--telemetry=network=on,page=off`
+  - `--chrome-policy <json>` - Custom Chrome enterprise policy as a JSON object. Kernel-managed policies (extensions, proxy, automation) are rejected server-side.
+  - `--chrome-policy-file <path>` - Read the Chrome enterprise policy from a file (use `-` for stdin). Mutually exclusive with `--chrome-policy`.
   - `--output json`, `-o json` - Output raw JSON object
   - _Note: When a pool is specified, omit other session configuration flags—pool settings determine profile, proxy, viewport, etc._
 - `kernel browsers delete <id-or-name>` - Delete a browser by ID or name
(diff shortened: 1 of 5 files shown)


masnwilliams (approved):
lgtm

PR #187, opened 2026-06-16
cli: show favicon and fix title on auth success page

pbjaru:
## Summary
The browser page shown after `kernel login` (`pkg/auth/success.html`) had two issues, both visible in the tab:

- **No favicon** — browsers fell back to the default globe icon. Embedded the Kernel favicon inline as an SVG data URI. Inlining (rather than serving `/favicon.ico`) avoids a race with the local callback server, which shuts down immediately after serving the page, so a follow-up request for the icon would fail.
- **Garbled title** — the `<title>` used an em-dash that rendered as mojibake (`â€"`) because the page declares no charset. Swapped it for a hyphen so the tab reads `kernel cli - authenticated`.

## Test plan
- [x] `go build ./...` passes
- [x] Verified the embedded data URI decodes to a valid Kernel mark SVG
- [x] Manual: run `kernel login`, confirm the browser tab shows the Kernel favicon and the title reads `kernel cli - authenticated` (both live in browser chrome, which headless screenshots can't capture)

---

> [!NOTE]
> **Low Risk**
> Cosmetic changes to the post-login HTML only; OAuth flow and token handling are unchanged.
> 
> **Overview**
> Improves the browser tab after **`kernel login`** by fixing the success page HTML and how it is assembled in **`pkg/auth/oauth.go`**.
> 
> Adds **`pkg/auth/favicon.svg`** and embeds it at build time. The callback handler still serves a single HTML response, but **`successHTML`** is now built from a template by substituting **`__KERNEL_FAVICON__`** with an inline **`data:image/svg+xml;base64,...`** URI so the tab icon works even though the local callback server shuts down right after the page is sent.
> 
> Updates **`success.html`**: a **`<link rel="icon">`** uses that placeholder, and the **`<title>`** switches from an em dash (which showed as mojibake without a charset) to a plain hyphen so the tab reads **`kernel cli - authenticated`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cc3a10d60910887f3c2382f96e41ff4789f29efe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


masnwilliams (approved):
lgtm

PR #194, opened 2026-07-09
Add refresh-on-profile-update as an optional param for browser pools

chruffins:
## Summary

Add `--refresh-on-profile-update` flag to `kernel browser-pools create` and `kernel browser-pools update`, plus bump `kernel-go-sdk` to v0.75.0 (which provides the `RefreshOnProfileUpdate` field on pool params and response types).

The new flag flushes idle browser instances when the pool's profile is updated. It's forwarded to the SDK as an explicit optional field (guarded by `BoolFlag.Set`) so omitting the flag on the command line leaves the SDK field unset and lets the server decide the default.

## Changes

- **`cmd/browser_pools.go`** — Add `RefreshOnProfileUpdate BoolFlag` to both `BrowserPoolsCreateInput` and `BrowserPoolsUpdateInput`, wire it through `Create()` / `Update()` to `params.RefreshOnProfileUpdate`, register the `--refresh-on-profile-update` CLI flag on both commands, and display the setting in `kernel browser-pools get` output.
- **`cmd/browser_pools_test.go`** — Add tests for:
  - Create with `RefreshOnProfileUpdate` set to `true`
  - Update with `RefreshOnProfileUpdate` set to `false`
  - Unset flag leaves the SDK field un-validated (server default)
  - `ProfileID` + `RefreshOnProfileUpdate` compose correctly when both are provided
- **`README.md`** — Document the new flag in the browser-pools reference section.
- **`go.mod` / `go.sum`** — Bump `kernel-go-sdk` from v0.72.0 to v0.75.0.

### DIFF
--- a/cmd/browser_pools.go
+++ b/cmd/browser_pools.go
@@ -91,14 +91,15 @@ func (c BrowserPoolsCmd) List(ctx context.Context, in BrowserPoolsListInput) err
 }
 
 type BrowserPoolsCreateInput struct {
-	Name             string
-	Size             int64
-	FillRate         int64
-	TimeoutSeconds   int64
-	Stealth          BoolFlag
-	Headless         BoolFlag
-	Kiosk            BoolFlag
-	ProfileID        string
+	Name                   string
+	Size                   int64
+	FillRate               int64
+	TimeoutSeconds         int64
+	Stealth                BoolFlag
+	Headless               BoolFlag
+	Kiosk                  BoolFlag
+	RefreshOnProfileUpdate BoolFlag
+	ProfileID              string
 	ProfileName      string
 	ProxyID          string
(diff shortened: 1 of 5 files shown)


cmd/browser_pools.go
```diff
@@ -139,6 +140,9 @@ func (c BrowserPoolsCmd) Create(ctx context.Context, in BrowserPoolsCreateInput)
 	if in.Kiosk.Set {
 		params.KioskMode = kernel.Bool(in.Kiosk.Value)
 	}
+	if in.RefreshOnProfileUpdate.Set {
```
hiroTamada, line 143: do you think we should have a check where profile has to be set when this flag is turned on?

hiroTamada (approved):

PR #201, opened 2026-07-13
Prevent OAuth login from hanging after callback

yummybomb:
## summary\n- use a bounded HTTP/1.1 client for OAuth token exchanges and refreshes\n- report token-exchange failures instead of leaving login waiting indefinitely\n- clarify that the browser page confirms callback receipt, not completed authentication\n\n## testing\n- go test ./...\n- go build ./cmd/kernel


yummybomb:
Closing in favor of further root-cause investigation.

PR #203, opened 2026-07-14
Set default browser pool fill rate

chruffins:
## Summary
address KERNEL-1593

- default `kernel browser-pools create` to a fill rate of 25% per minute
- add coverage that verifies the default is sent in the create request

## Test plan
- [x] `go test ./cmd`
- [x] `make build` and create a browser pool, then verify the fill rate is 25% per minute


hiroTamada (approved):

hiroTamada (commented):
lets make the default value in the backend

PR #208, opened 2026-07-20
Refresh brew tap before upgrading in kernel upgrade

masnwilliams:
## Summary

`kernel upgrade` detects a newer version from GitHub releases, then shells out to `brew upgrade kernel/tap/kernel`. But `brew upgrade` doesn't refresh the tap on its own, and Homebrew throttles its auto-update — so right after a release the local tap is stale, brew still sees the old formula, and the upgrade silently no-ops with "already installed" (exit 0). The user sees `kernel upgrade` announce a new version and then nothing changes.

Repro: release a new version, then on a machine whose tap hasn't refreshed run `kernel upgrade` → "New version available: X → Y" followed by "Warning: kernel/tap/kernel X already installed", no upgrade.

## Fix

- `kernel upgrade` now runs `brew update` before `brew upgrade kernel/tap/kernel` (brew method only; npm/pnpm/bun already pull `@latest`).
- `executeUpgrade`/`upgradeCommandArgs` refactored to run an ordered list of commands so the brew path can do two steps; stderr from all steps is still captured for the existing old-tap error diagnosis.
- The passive suggested-upgrade hint now shows `brew update && brew upgrade kernel/tap/kernel` for the same reason.

## Testing

`go build ./...`, `go vet`, `gofmt`, and `go test ./pkg/update/...` all pass (suggestion-string test updated).

---

> [!NOTE]
> **Low Risk**
> CLI-only change to upgrade command sequencing and user-facing hints; no auth, data, or core runtime behavior beyond fixing stale-tap brew upgrades.
> 
> **Overview**
> Fixes **Homebrew upgrades silently doing nothing** right after a release when the local tap is stale: `kernel upgrade` now runs **`brew update` before `brew upgrade kernel/tap/kernel`** for the brew install path only.
> 
> Upgrade execution is refactored so each install method returns an **ordered list of shell commands** (`[][]string`). `executeUpgrade` runs them sequentially and still aggregates **stderr** across steps for the existing old-tap error handling. Dry-run / displayed commands use **`brew update && brew upgrade kernel/tap/kernel`**.
> 
> Passive upgrade hints in **`SuggestUpgradeCommand`** (brew and unknown-method fallback) use the same two-step brew string. Tests cover brew command ordering and the updated suggestion text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 13320545f5f8cfa8c0570398eed9366c467615ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -98,18 +98,24 @@ func isOldTapError(stderr string) bool {
 	return strings.Contains(stderr, "brew tap kernel/tap")
 }
 
-// upgradeCommandArgs returns the command and arguments for a given installation method.
-// Returns nil if the method is unknown.
-func upgradeCommandArgs(method update.InstallMethod) []string {
+// upgradeCommandArgs returns the ordered commands to run for a given installation
+// method. Returns nil if the method is unknown.
+func upgradeCommandArgs(method update.InstallMethod) [][]string {
 	switch method {
 	case update.InstallMethodBrew:
-		return []string{"brew", "upgrade", "kernel/tap/kernel"}
+		// brew upgrade doesn't refresh the tap first, so a freshly released
+		// formula is invisible and the upgrade silently no-ops ("already
+		// installed"). Update the tap before upgrading.
+		return [][]string{
+			{"brew", "update"},
+			{"brew", "upgrade", "kernel/tap/kernel"},
+		}
 	case update.InstallMethodPNPM:
-		return []string{"pnpm", "add", "-g", "@onkernel/cli@latest"}
+		return [][]string{{"pnpm", "add", "-g", "@onkernel/cli@latest"}}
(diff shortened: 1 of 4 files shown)


pbjaru (approved):
LGTM

non blocking: for `cmd/upgrade.go:103` - could we add a small unit test confirming brew update runs before brew upgrade?

masnwilliams:
Added `cmd/upgrade_test.go` — `TestUpgradeCommandArgsBrewUpdatesBeforeUpgrade` asserts the `brew update` step precedes `brew upgrade` for the brew method, plus table coverage of the per-method command args and the display string. `go test ./cmd/...` green.

PR #226, opened 2026-08-12
Add private host network flags

hiroTamada:
## Summary

- bump the Go SDK to v0.89.0 for browser network configuration types
- add `--private-host` to browser and browser-pool creation
- add `--clear-private-hosts` to restore browser-pool defaults after an override
- preserve the API's omitted and configured network semantics without exposing the niche empty-list control in the CLI

## Testing

- `make test`
- production smoke: browser create/get/delete with `--private-host`
- production smoke: pool create/get, replace, clear, and delete lifecycle
- verified `--disable-private-hosts` is not exposed by browser or browser-pool commands


yummybomb (approved):
no blocking findings. one non-blocking comment worth leaving:

does it make sense to surface network.private_hosts in the human-readable browsers get and browser-pools get output? currently users can configure or clear it, but can only verify the effective setting with --output json

PR #232, opened 2026-08-17
Add org entitlements command

IlyaasK:
## Why

The organization entitlements API and Go SDK v0.92.0 are released. The CLI should expose the effective organization policy calculated by the API without maintaining a second plan-to-entitlement table.

## What

- upgrade kernel-go-sdk from v0.91.0 to v0.92.0
- add the read-only `kernel org entitlements` command
- show effective plan, feature access, feature constraints, and organization limits in human-readable output
- support `--output json` using the generated SDK model unchanged
- render nullable caps as `unlimited` in human output while preserving JSON `null`

`kernel org limits` remains the command for reading and changing configurable organization concurrency limits.

## How

The command calls the generated `Organization.Entitlements.Get` SDK operation directly. All entitlement resolution remains in the API; the CLI only projects the returned model.

## Verification

Before: kernel-go-sdk v0.91.0 and the CLI did not expose organization entitlements.

After:

- `make test` passes (`go vet ./...` and `go test ./...`)
- `make build` passes
- focused organization command tests pass
- command registration and help output verified locally
- read-only production smoke passes for both human and JSON output
- production JSON was structurally asserted for plan, trial, managed-auth, extension-cap, and browser-limit fields without printing organization data

KERNEL-1713

---

> [!NOTE]
> **Low Risk**
> Read-only CLI surface over a new SDK GET; no auth, persistence, or limit-mutation behavior changes.
> 
> **Overview**
> Adds a read-only **`kernel org entitlements`** command that surfaces the API’s effective organization policy (plan, trial, features, and limits) without duplicating entitlement logic in the CLI.
> 
> The command calls **`Organization.Entitlements.Get`** from **kernel-go-sdk v0.92.0** (upgraded from v0.91.0). Human output is a table projection of the SDK model; **`--output json`** prints the generated model as-is. Nullable caps render as **`unlimited`** / **`none`** / **`unknown`** in the table while JSON keeps **`null`** where applicable.
> 
> README documents the new subcommand. **`kernel org limits`** is unchanged for reading and updating configurable concurrency settings.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4961b765399465608c246189bb015c031643058a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/org.go
+++ b/cmd/org.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/kernel/cli/pkg/util"
 	"github.com/kernel/kernel-go-sdk"
@@ -19,8 +20,14 @@ type OrgLimitsService interface {
 	Update(ctx context.Context, body kernel.OrganizationLimitUpdateParams, opts ...option.RequestOption) (res *kernel.OrgLimits, err error)
 }
 
+// OrgEntitlementsService defines the organization entitlements operation used by the CLI.
+type OrgEntitlementsService interface {
+	Get(ctx context.Context, opts ...option.RequestOption) (res *kernel.OrgEntitlements, err error)
+}
+
 type OrgCmd struct {
-	limits OrgLimitsService
+	limits       OrgLimitsService
+	entitlements OrgEntitlementsService
 }
 
 type OrgLimitsGetInput struct {
@@ -30,6 +37,32 @@ type OrgLimitsGetInput struct {
 type OrgLimitsSetInput struct {
 	DefaultProjectMaxConcurrentSessions Int64Flag
 	Output                              string
+}
+
+type OrgEntitlementsInput struct {
+	Output string
+}
+
+func (c OrgCmd) Entitlements(ctx context.Context, in OrgEntitlementsInput) error {
+	if err := validateJSONOutput(in.Output); err != nil {
+		return err
+	}
+
+	entitlements, err := c.entitlements.Get(ctx)
+	if err != nil {
+		return util.CleanedUpSdkError{Err: err}
+	}
+
+	if in.Output == "json" {
+		if entitlements == nil {
+			fmt.Println("null")
+			return nil
+		}
+		return util.PrintPrettyJSON(entitlements)
+	}
+
+	renderOrgEntitlements(entitlements)
+	return nil
 }
 
 func (c OrgCmd) LimitsGet(ctx context.Context, in OrgLimitsGetInput) error {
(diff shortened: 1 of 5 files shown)


cmd/org.go
```diff
@@ -125,6 +158,56 @@ func orgLimitFieldPresent(field respjson.Field) bool {
...
+	}
+
+	status := entitlements.Plan.Status
+	if !entitlements.Plan.JSON.Status.Valid() {
+		status = "none"
+	}
+	trialEndsAt := "none"
+	if entitlements.Plan.JSON.TrialEndsAt.Valid() {
+		trialEndsAt = entitlements.Plan.TrialEndsAt.Format(time.RFC3339)
+	}
```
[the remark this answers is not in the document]
IlyaasK: Fixed in c820e44. Human-readable trial end timestamps now use the shared util.FormatLocal formatter, matching the rest of the CLI, while JSON output remains unchanged. Added a focused regression test for a non-null trial timestamp; focused org tests and the full make test suite pass.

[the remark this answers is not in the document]
IlyaasK: resolved

cmd/org.go
```diff
@@ -125,6 +157,56 @@ func orgLimitFieldPresent(field respjson.Field) bool {
...
 }
 
+func renderOrgEntitlements(entitlements *kernel.OrgEntitlements) {
+	if entitlements == nil {
+		pterm.Info.Println("No organization entitlements found")
+		return
+	}
+
+	status := entitlements.Plan.Status
+	if !entitlements.Plan.JSON.Status.Valid() {
```
masnwilliams: **[P2] distinguish explicit null from an invalid required response**

`Field.Valid()` is false not only for explicit JSON `null`, but also when the field is omitted or fails to decode. The same issue applies to `TrialEndsAt` and to the nullable caps passed through `formatProjectLimitValue` below. That means an incomplete or malformed API response is rendered as `none`/`unlimited`, which can overstate the organization's effective policy instead of exposing a contract failure. Please branch on `field.Raw() == respjson.Null` for the explicit-null meaning and treat omitted/malformed required fields as `unknown` or return an error. Tests should cover populated, explicit-null, omitted, and malformed inputs.

cmd/org_test.go
```diff
@@ -16,6 +18,101 @@ type FakeOrgLimitsService struct {
...
+	buf := capturePtermOutput(t)
+	c := OrgCmd{entitlements: &FakeOrgEntitlementsService{
+		GetFunc: func(ctx context.Context, opts ...option.RequestOption) (*kernel.OrgEntitlements, error) {
+			return testOrgEntitlements(t), nil
+		},
+	}}
+
+	assert.NoError(t, c.Entitlements(context.Background(), OrgEntitlementsInput{}))
+	out := buf.String()
+	assert.Contains(t, out, "Contractual plan")
```
masnwilliams: **[P2] verify the complete entitlement projection**

This renderer manually maps roughly 27 SDK fields into rows, but these substring assertions only prove that a few labels and one shared value appear somewhere. A dropped row, swapped field, incorrect boolean, or duplicated value would still pass. Please extract the row construction into a pure `orgEntitlementRows` helper (or capture an equivalent structured result) and compare the complete `pterm.TableData`/golden output. Give every source field a distinct value so the test protects the full projection as the SDK model evolves.

masnwilliams (changes requested):
requesting changes on two boundary/testability issues. the direct SDK-backed command and flattened table are otherwise the right shape; this does not need a larger architectural rewrite.

cmd/org_test.go
IlyaasK: Fixed in dc98465. I extracted `orgEntitlementRows` as the pure projection boundary and replaced the substring checks with an exact `pterm.TableData` assertion covering the header and every plan, feature, and limit row. The fixture uses distinct plan/numeric values and mixed booleans, while the existing explicit-null/unlimited and local-time tests remain separate. `GOCACHE=/private/tmp/kernel-cli-pr-232-go-cache make test` passes (go vet plus all packages).

cmd/org.go
IlyaasK: Fixed in 444eae7. Human entitlement rendering now treats only `field.Raw() == respjson.Null` as `none`/`unlimited`; omitted and malformed required fields render `unknown`. Nullable status also validates that the raw JSON is actually a string because the SDK accepts a numeric enum value as valid. Added table-driven coverage for populated, explicit-null, omitted, and malformed status, trial end, max stored extensions, and max auth connections. Full `make test` passes (go vet plus all packages).

masnwilliams (approved):
both requested changes are addressed: nullable response states are distinguished correctly, and the full entitlement projection now has structured coverage with boolean provenance checks. local vet, test, build, and diff checks pass on 4961b76.

PR #234, opened 2026-08-23
Add fx MCP install target

masnwilliams:
## Summary

- add `fx` as a `kernel mcp install` target
- write Kernel's Streamable HTTP and OAuth configuration to `~/.fx/mcp.json`
- preserve existing fx MCP servers and secure the profile directory and file
- print fx-specific authentication and verification steps

## Testing

- `make test`
- built the CLI and smoke-tested `kernel mcp install --target fx` with a clean home directory
- verified generated permissions are `0700` for `~/.fx` and `0600` for `mcp.json`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Local CLI config-file writes only; no auth, network, or shared-service changes. Restrictive file modes slightly reduce risk of leaking MCP config.
> 
> **Overview**
> Adds `fx` as a `kernel mcp install` target so Kernel can be registered as an HTTP MCP server with OAuth in `~/.fx/mcp.json`.
> 
> Install merges a `kernel` entry into the existing `mcp` map without dropping other servers, writes the file with `0600` (and `0700` for a newly created `~/.fx`), and prints fx-specific reload/auth/verify steps. Tests cover merge behavior and permission handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e501883bbc52c02b948cac319e601073966eb15. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/mcp/install.go
+++ b/cmd/mcp/install.go
@@ -24,6 +24,7 @@ Supported targets:
   vscode      - Visual Studio Code
   goose       - Goose AI
   zed         - Zed editor
+  fx          - fx coding agent
 
 Examples:
   # Install for Cursor
@@ -135,5 +136,11 @@ func printPostInstallInstructions(target Target) {
 		pterm.Println("  1. Restart Zed")
 		pterm.Println("  2. The Kernel context server will be available")
 		pterm.Println("  3. You'll be prompted to authenticate when first using Kernel tools")
+
+	case TargetFx:
+		pterm.Info.Println("Next steps:")
+		pterm.Println("  1. Start fx, or run '/mcp reload' in an existing session")
+		pterm.Println("  2. Run '/mcp auth kernel --open' to authenticate")
+		pterm.Println("  3. Run '/mcp list' to verify that Kernel is connected")
 	}
 }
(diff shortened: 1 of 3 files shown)


rgarcia (approved):
reviewed — looks good, approving. one suggestion worth considering before/after merge:

### Suggestions
- `cmd/mcp/mcp.go:248-250` — the unconditional `Chmod(dir, 0700)` repermissions a pre-existing `~/.fx` that holds unrelated profile state (and on non-macOS, the credential store). Security value is near-zero once the file itself is 0600 — a world-readable dir only leaks filenames. Consider dropping it so existing dirs keep their permissions; `MkdirAll(dir, 0700)` already covers the create-new case.

### Nits
- `cmd/mcp/mcp_test.go:17` / `cmd/mcp/mcp_test.go:82-87` — the test pre-creates the dir at 0755 and then asserts 0700, which currently *requires* the unconditional chmod above. Update alongside any change to `writePrivateJSONFile`.
- `cmd/mcp/mcp_test.go` — no coverage for the clean-install path (no pre-existing file/dir); a second small case would lock in create-at-0700/0600 behavior.

### Follow-up (not this PR)
- other targets' configs also hold tokens (`~/.claude.json` especially) — worth a follow-up ticket to apply private-perm writes there

Also verified the fx integration against fx's published MCP docs (fx.sh/docs/capabilities/mcp): config path/schema, `"type": "http"` + `url`, empty `oauth` object selecting delegated user authorization, and all three `/mcp` commands in the post-install copy match exactly.

PR #248, opened 2026-09-10
Add vault provider configuration commands and wallet imports

rgarcia:
## Summary

- Add `vault-provider-configs create/list/get/show/update/delete` through the generated SDK service. Support bounded pagination, safe JSON/table output, rename/secret-only rotation, confirmation, and unused-config deletion conflicts. Missing configurations remain errors on delete; existing vault/item deletion compatibility is unchanged.
- Add immutable wallet selection by `--provider-config-id` or `--provider-config-name`, preserving existing Kernel-managed specs. AgentCard uses configured enrollment/reuse; Link imports an existing backend OAuth grant through `--tokens-file`.
- Accept client credentials and grants only through protected files or stdin. Reject tokens/credentials at `--spec` protocol boundaries without restricting opaque metadata; filter write-only/unknown response fields and withhold credential-bearing API/transport error details even through root error formatting. No automatic SDK retries.
- Display resolved config bindings and recovery event operations. Treat `recovery_required` as unresolved rather than declined/expired, refuse stale invocation/open actions, and explain imported-grant degradation without suggesting in-place replacement or retrying uncertain payments. Correct pending card-update omission/empty-list guidance. Use one effective-action policy for execution and human output while preserving API-advertised fields in display-safe JSON.
- Share offset-pagination parsing across projects, vaults, and provider configurations, accepting omitted terminal offsets and rejecting non-advancing continuation offsets.

## Official SDK dependency

Pins the published `github.com/kernel/kernel-go-sdk v0.101.0`, release commit **`68050bf6a52dfa65e4654966e728cf8ee53d8f8a`**. The SDK release prerequisite is resolved, and the temporary preview-only instructions have been removed. No staging dependency, replacement, or alternate modfile is required.

The official release's `vaultproviderconfig.go` and `vaultitem.go` are byte-identical to the previously tested preview commit `87471646cd377f2e3503cd98f69f233fe93f479c`. Config CRUD, request-only wallet tokens, config references, immutable updates, and recovery statuses match the inspected API contract. The generated config methods are `New`, `Get`, `List`, `Update`, and `Delete`; no raw HTTP fallback is used.

## Validation

All passing with **the official v0.101.0 dependency and the normal committed modfile**, Go 1.25.0:

- `go test -count=1 ./...`
- `go test -race -count=1 ./cmd -run 'Vault|OffsetPagination|ProjectsList'`
- `go build ./...`
- `go vet ./...`
- Built the executable and inspected config/wallet command help.

Focused tests cover both providers, file/stdin input, config CRUD wire shapes, omitted vs rotated credentials, immutable-field rejection, config selection conflicts, imported token mapping, secret-safe output/errors, pagination including an omitted final-page offset, 403/404/409/429/500 without retry, old default flows, recovery behavior, and pending-update field presence. Regressions also cover stale recovery actions in list/get output, legitimate credential-named metadata, raw imported-authorization field preservation, and shared pagination behavior across all three callers. New credential sentinels are generated at runtime rather than stored in fixtures.

No live provider calls or deployments were performed during validation.


cmd/vault_provider_configs.go
```diff
@@ -0,0 +1,233 @@
...
+					pterm.Info.Println("Deletion cancelled")
+					return nil
+				}
+			}
+			err := c.configs.Delete(cmd.Context(), args[0], option.WithMaxRetries(0))
+			if err != nil && !util.IsNotFound(err) {
+				return vaultCredentialError(err)
+			}
+			pterm.Success.Println("Deleted or not found: vault provider configuration " + args[0])
+			return nil
```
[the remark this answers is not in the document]
rgarcia: Fixed in e4c5b68: configuration deletion now reports HTTP 404 as a sanitized error and prints success only after HTTP 204. The existing vault/item 404-as-success behavior is unchanged. Updated the delete-response regression matrix; full tests, focused vault race tests, build, and vet pass with the exact preview SDK replacement.

cmd/vaults_output.go
```diff
@@ -271,6 +288,13 @@ func printVaultItem(item *kernel.VaultItemUnion, output string) error {
 		}
 	}
 	PrintTableNoPad(rows, true)
+	if item.State.Status == "recovery_required" {
+		pterm.Warning.Println("recovery_required: the original operation is unresolved, not declined or expired. Do not retry, delete, or replace it. Reconcile with the provider or support; no reset operation exists.")
```
hiroTamada, line 283: recovery policy is scattered across invocation, opening, operation hints, and rendering, and the paths already disagree. this guard runs after the property table: a recovery item still prints `Required action: spend_approval`. `items list` advertises the stale action too; reproduced both locally.

can we derive effective actions/operations once and use that policy for execution and human output, keeping raw JSON preservation explicit? adding more status checks at individual rendering sites will make this harder to maintain. `printVaultItem` cyclomatic complexity rises from 28 to 37 in this change.

cmd/vaults_secrets.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+	return values, nil
+}
+
+func vaultSpecHasSecrets(value json.RawMessage) bool {
+	var object map[string]json.RawMessage
+	if json.Unmarshal(value, &object) == nil {
+		for key, child := range object {
+			switch strings.ToLower(key) {
```
hiroTamada: this recursive field-name blacklist also runs through the shared card parser and rejects arbitrary merchant metadata. a non-secret `metadata: {"tokens":"loyalty-points"}` now fails before reaching the API; reproduced locally. field names alone don't establish that a value is a credential.

can we enforce credential restrictions at explicit protocol boundaries without reserving these words inside opaque metadata, and add a regression test for legitimate metadata?

cmd/vaults_wallet_spec.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if err := validateVaultConfigReference(auth.Client.ProviderConfig); err != nil {
+		return err
+	}
+	tokens, err := readVaultSecrets(cmd, "tokens-file", "access_token", "refresh_token")
+	if err != nil {
+		return err
+	}
+	var authorization map[string]json.RawMessage
+	_ = json.Unmarshal(spec["authorization"], &authorization)
```
hiroTamada: `configureVaultWallet` mixes flag parsing, validation, file I/O, and JSON mutation (cyclomatic complexity 19). authorization is constructed using nested `map[string]any`, decoded into an anonymous struct, then decoded again here into a map for token insertion. the SDK-typed return still uses `param.Override`, so that boundary doesn't validate the assembled shape.

can we normalize selection flags into one reference, separate credential reading from assembly, and assemble authorization once with an explicit shape? preserve untouched raw spec fields and omission semantics; this should simplify the direct flow rather than introduce a generic builder framework.

cmd/vault_provider_configs.go
```diff
@@ -0,0 +1,233 @@
...
+func (c VaultProviderConfigsCmd) List(ctx context.Context, limit, offset int64, output string) error {
+	if limit < 1 || limit > 100 || offset < 0 {
+		return fmt.Errorf("--limit must be between 1 and 100; --offset must be non-negative")
+	}
+	var response *http.Response
+	page, err := c.configs.List(ctx, kernel.VaultProviderConfigListParams{Limit: kernel.Opt(limit), Offset: kernel.Opt(offset)}, option.WithMaxRetries(0), option.WithResponseInto(&response))
+	if err != nil {
+		return vaultCredentialError(err)
+	}
+	hasMore, err := strconv.ParseBool(response.Header.Get("X-Has-More"))
```
hiroTamada: vault listing already uses `parseProjectListPagination`. this duplicates the same header protocol while adding useful handling for omitted final-page offsets and non-advancing offsets, leaving the existing parser with different behavior.

can we extract shared offset-pagination parsing, incorporate these checks, and migrate both callers? keep credential-safe error formatting at the command boundary.

hiroTamada (approved):
approving with the following non-blocking comments on recovery policy, credential validation boundaries, wallet request assembly, and shared pagination parsing.

validation: full tests, focused vault race tests, and go vet ./... pass. additional local probes confirmed the stale recovery-action display and rejection of non-secret metadata. no changed Go file crosses 1,000 lines.

cmd/vaults_output.go
rgarcia: Addressed in ba5a9f7. Execution, browser opening, list/get human output, and operation hints now share effectiveVaultItemActions. Recovery suppresses the required action, action/approval URLs, and operations; display-safe JSON deliberately preserves the API-advertised fields. Added get/list regressions for requested versus recovery states in human and JSON output, and strengthened the stale-action opening test. Split rendering guidance out of the property-table function.

cmd/vaults_secrets.go
rgarcia: Addressed in ba5a9f7. Credential checks now traverse only protocol objects (authorization, client, and provider_config), not arbitrary objects or arrays. Opaque metadata can use these field names. Added both-provider regressions across wallet create and card create/update, while retaining rejection tests for credential fields at the actual protocol boundaries.

cmd/vaults_wallet_spec.go
rgarcia: Addressed in ba5a9f7. Selection flags normalize into one reference, provider-specific validation is separate, and authorization/client/token shapes are explicit structs. Existing authorization is parsed once into retained raw fields, with only method/client decoded for validation; token-file reading stays separate from assembly. Added omission/null/empty-list/false/large-integer and unknown-client-field preservation tests. The final SDK override remains intentional to preserve raw spec extensions, rather than being treated as validation.

cmd/vault_provider_configs.go
rgarcia: Addressed in ba5a9f7. Projects, vaults, and provider configurations now use parseOffsetPagination with the current offset. The shared parser accepts an omitted final-page offset and rejects missing, malformed, overflowing, negative, or non-advancing continuation offsets. Moved and expanded parser tests and added all-three-caller regressions. Vault/config commands still sanitize pagination errors at their boundaries. Full tests, focused vault/pagination/projects race tests, build, and vet pass.

PR #254, opened 2026-09-16
Preserve public credential values and clarify fill errors

rgarcia:
## Summary
- Return explicitly non-sensitive text/email values in credential JSON output across create, update, get, list, and collect. Sensitive values, TOTP seeds, missing sensitivity flags, and unknown fields remain protected. Preserve collection URLs.
- Give fill API errors actionable explanations alongside HTTP status and allowlisted codes. Documented pre-write rejections (400/403/404/409) state that this request wrote no fields; transport loss and other uncertain failures retain no-retry guidance. Raw upstream details remain hidden.
- Let the API validate type-specific fields, formats, destinations, and selector limits rather than duplicating policy in the CLI. Retain structural JSON validation, size limits, advertised-operation checks, safe result projection, and zero SDK retries.
- Update help/README and add public-value and meaningful-error regression coverage.

## Validation
- `make test` (go vet and all tests): passed
- `make build`: passed
- Focused vault public-value and fill-error tests under the race detector: passed
- Existing subprocess tests still verify one attempt and safe stdout/stderr behavior for errors and uncertain outcomes.
- No live API/browser/payment operations performed. No release included.


masnwilliams (approved):
approved. the public-value projection is fail-closed and the tests cover sensitive, TOTP, unknown, and missing-metadata cases. delegating item-specific fill policy to the API removes duplicated validation, and the new guidance matches the API's pre-write error contract.

non-blocking follow-up: preflight item retrieval failures still return the generic `could not retrieve vault item` message, so missing vault/item errors do not benefit from the new status/code guidance.

PR #260, opened 2026-09-17
Handle OAuth denial and deduplicate login output

rgarcia:
## Summary

OAuth consent denial returns `error=access_denied` without a code. The CLI previously treated that valid error response as a malformed success callback, showing “missing authorization code” in the browser. Native login also emitted duplicate terminal statuses and reported success before credentials were persisted.

- Handle OAuth errors after state validation and before requiring a code. Denial returns a distinct error without entering token exchange.
- Reuse the callback page layout for a clear branded denial page without a success checkmark. Provider-supplied error descriptions are not reflected.
- Stop and erase the waiting spinner without printing a second status. The root renderer now emits the single meaningful denial, cancellation, or authentication error with a nonzero exit.
- Classify cancellation from the authentication error itself so a racing context cancellation cannot mask an unrelated failure.
- Print one success only after credentials are saved. Credential-save failures retain the storage error and reauthentication guidance and now exit nonzero.
- Preserve existing stored credentials on denial.
- Extract the callback and root error handlers for focused rendered-output regression coverage.

## Validation

- `go test ./pkg/auth ./cmd -count=1`
- `go test -race ./pkg/auth -count=1`
- `go vet ./pkg/auth ./cmd`
- `go build ./...`
- Focused output tests cover success, consent denial, generic authentication failure, cancellation races, credential-save failure, and the real spinner erase boundary.
- Earlier built-binary loopback fixture: simulated browser launch and denial callback produced one clear terminal error, exit code 1, zero token requests, and unchanged fixture credentials. This used isolated fixture configuration with keyring access disabled, not a live provider login.
- Chromium rendering at desktop and mobile widths: denial text, embedded favicon, no success checkmark, and no horizontal overflow. Existing callback layout and colors are retained.
- Build, test, Semgrep, Socket Security, and BugBot checks pass on `d99e4da9b708e119211d76dec6e5047305d8858f`.

Released as `v0.39.1` from merge commit `76daab28eabfc97bdb729d3fa579d9e2bbe99ecb`. A live provider denial retest is still pending; the loopback fixture above was local only.


cmd/login.go
```diff
@@ -55,32 +56,37 @@ func runLogin(cmd *cobra.Command, args []string) error {
...
+}
 
+func completeLogin(
+	ctx context.Context,
+	spinner spinnerStopper,
+	authenticate func(context.Context) (*auth.TokenStorage, error),
+	saveTokens func(*auth.TokenStorage) error,
+) error {
+	tokens, err := authenticate(ctx)
+	_ = spinner.Stop()
```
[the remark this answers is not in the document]
rgarcia: fixed in d45fff2: the login spinner now uses RemoveWhenDone, so Stop clears the waiting frame before the single final success or root-rendered error.

cmd/login.go
```diff
@@ -55,32 +56,37 @@ func runLogin(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		if errors.Is(err, auth.ErrAuthorizationDenied) {
+			return err
+		}
+		if errors.Is(ctx.Err(), context.Canceled) {
```
masnwilliams: non-blocking: can we classify this from the returned error (`errors.Is(err, context.Canceled)`) rather than the ambient context state? if cancellation races with a different auth failure, `ctx.Err()` can mask the real error as a user cancellation. small edge case, so not worth blocking this PR.

cmd/login_test.go
```diff
@@ -0,0 +1,142 @@
...
+				renderCommandError(output, fang.Styles{
+					ErrorText: lipgloss.NewStyle(),
+					Program:   fang.Program{Flag: lipgloss.NewStyle()},
+				}, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			rendered := ansiEscapes.ReplaceAllString(output.String(), "")
+			assert.Equal(t, 1, spinner.stops)
```
masnwilliams, line 144: non-blocking: this fake proves that `Stop` is called, but it cannot catch a regression in the actual spinner configuration—the earlier version without `WithRemoveWhenDone` would still pass. worth adding a narrow test around the real spinner/output boundary when convenient.

masnwilliams (approved):
approved. this is a focused improvement: OAuth denial is handled after state validation, terminal output now has one owner, success is reported only after credentials are persisted, and the new failure paths have solid coverage. the callback refactor also leaves the files easier to navigate.

left two non-blocking follow-ups around cancellation classification and coverage of the real spinner behavior.

cmd/login_test.go
rgarcia: addressed in d99e4da: the production spinner construction now goes through newLoginSpinner, and TestLoginSpinnerClearsWaitingOutput drives that real pterm spinner against a terminal-output buffer and verifies Stop erases the waiting line. I also removed WithRemoveWhenDone locally to prove this regression test fails, then restored it; focused and package tests pass.

cmd/login.go
rgarcia: addressed in d99e4da: cancellation is now classified with errors.Is(err, context.Canceled), and the output test covers an already-canceled ambient context returning an unrelated callback failure to ensure the real failure remains visible. Verified with go test ./cmd -run 'Test(CompleteLoginOutput|LoginSpinnerClearsWaitingOutput)$' -count=1.

rgarcia:
bugbot run

